PR #8, opened 2024-05-03
DRAFT: Propose an alternative to the types in capacity_types.go

klueska:
I spent some time today going through the types in the file `pkg/api/capacity_types.go` in detail. I have proposed my changes to it as an entirely new file (instead of a diff to the original) so that we can more easily comment on it line-by-line without the diff getting in the way.

**The biggest thing to note is the following:**
I'm coming around to the idea that `DevicePool` (or whatever future resource models we come up with) _**could**_ be 
top level objects, rather than wrapping them in multiple-levels of indirection like we have today to support the `NamedResources` model:
```
type ResourceSlice struct {
	metav1.TypeMeta
	metav1.ObjectMeta

	NodeName string `json:"nodeName,omitempty" protobuf:"bytes,2,opt,name=nodeName"`
	DriverName string `json:"driverName" protobuf:"bytes,3,name=driverName"`
	ResourceModel `json:",inline" protobuf:"bytes,4,name=resourceModel"`
}

type ResourceModel struct {
	NamedResources *NamedResourcesResources
}

type NamedResourcesResources struct {
	Instances []NamedResourcesInstance `json:"instances" protobuf:"bytes,1,name=instances"`
}
```

Having all of these levels of indirection doesn't actually add much value. The original idea was that the scheduler (or whoever else cares about `ResourceSlices`) only has one object type to subscribe to, and we can simply extend its embedded `ResourceModel` with more types over time.

This actually makes it awkward, however, when we need to split our `Instances` at the lowest level across multiple API server objects. Each object needs to be wrapped in multiple levels of indirection, making it a somewhat hard to reason about when trying to reconstruct all objects from a given driver.

On the contrary, by making `DevicePool` (or whatever future resource models we come up with) a top-level API server object, we have a natural way of breaking things apart. Each pool is it's own self-contained object, so there is no ambiguity on where the object boundaries should be. After thinking about it for a while, this actually feels like the more natural and "Kubernetes-like" way of abstracting these objects.

The major difference being that now the scheduler can't just subscribe to a single `ResourceSlices` object and have all supported model implementations from all drivers follow that. It potentially has to subscribe to multiple top-level objects representing different models and react to them accordingly. This seems like a reasonable trade-off since one would always have to write the code to be aware of the different models anyway.

**The second biggest thing to note is the following:**
I have (slightly) renamed and "future-proofed" all of the types that represent named resources and their requests. Whether we want to do this is still being debated, but I wanted to get a concrete proposal for what it would look like as a diff to @johnbelamaric's prototype, rather than continuing to talk in "what-ifs".

**The third biggest thing to note is the following:**
We can now have a meaningful `Status` attached to the top-level type. In this case I have expanded it to include all allocated devices and the claims they have been allocated to.

### DIFF
--- a/k8srm-prototype/pkg/api/capacity_types_alternate.go
+++ b/k8srm-prototype/pkg/api/capacity_types_alternate.go
@@ -1,0 +1,177 @@
+package api
+
+import (
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	DevMgmtAPIVersion = "devmgmtproto.k8s.io/v1alpha1"
+)
+
+// DevicePool represents a collection of devices managed by a given driver. How
+// devices are divided into pools is driver-specific, but typically the
+// expectation would a be a pool per identical collection of devices, per node.
+// It is fine to have more than one pool for a given node, for the same driver.
+type DevicePool struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   DevicePoolSpec   `json:"spec,omitempty"`
+	Status DevicePoolStatus `json:"status,omitempty"`
+}
+
+// DevicePoolSpec identifies the driver and contains the data for the pool
+// prior to any allocations.
+// NOTE: It's not clear that spec/status is the right model for this data.
+type DevicePoolSpec struct {
+	// NodeName is the name of the node containing the devices in the pool.
+	// For network attached devices, this may be empty.
+	// +optional
+	NodeName *string `json:"nodeName,omitempty"`
+
+	// Driver is the name of the DeviceDriver that created this object and
+	// owns the data in it.
+	//
+	// +required
+	DriverName string `json:"driverName"`
+
+	// Attributes contains device attributes that are common to all devices
+	// in the pool.
+	//
+	// +optional
+	Attributes []Attribute `json:"attributes,omitempty"`
+
+	// SharedResources are resources that are shared by all devices in the
+	// pool. This is typically used when representing a partitionable
+	// device, and need not be populated otherwise.
+	//
+	// +optional
+	SharedResources []DeviceResource `json:"sharedResources,omitempty"`
+
+	// Devices contains the individual devices in the pool.
+	//
+	// +required
+	Devices []Device `json:"devices"`
+}
+
+// DevicePoolStatus contains the state of the pool as last reported by the
+// driver. Note that this will not include the allocations that have been made
+// by the scheduler but not yet seen by the driver. Thus, it is NOT sufficient
+// to make future scheduling decisions.
+type DevicePoolStatus struct {
+	AllocatedDevices []AllocatedDevice `json:"allocatedDevices,omitempty"`
+}
+
+// AllocatedDevice represents a device that has been allocated from the pool.
+type AllocatedDevice struct {
+	Name string
+	ClaimUID types.UID
+}
+
+// Device is used to track individual devices in a pool.
+type Device struct {
+	// Name is a driver-specific identifier for the device.
+	//
+	// +required
+	Name string `json:"name"`
+
+	// Attributes contain additional metadata that can be used in
+	// constraints. If an attribute name overlaps with the pool attribute,
+	// the device attribute takes precedence.
+	//
+	// +optional
+	Attributes []Attribute `json:"attributes,omitempty"`
+
+	// SharedResourceRequests contains requests for some subset of the shared
+	// resources available in the overall pool. They represent the set of
+	// shared resources consumed by this device when it gets allocated.
+	//
+	// +optional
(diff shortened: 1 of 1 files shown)


pohly:
> The original idea was that the scheduler (or whoever else cares about ResourceSlices) only has one object type to subscribe to, and we can simply extend its embedded ResourceModel with more types over time.

This is what https://github.com/kubernetes/kubernetes/issues/123699 depends upon: kubelet knows that drivers may access ResourceSlice and ResourceClaim through the REST proxy and denies access to everything else. When combining kubelet from 1.31 with a future DRA driver which wants to publish a "SomeFutureType" instead of a "ResourceSlice", the driver's request will get denied. I'm working on a KEP update for this issue and have a prototype implementation ready that I'll share together with it.

> This actually makes it awkward, however, when we need to split our Instances at the lowest level across multiple API server objects. Each object needs to be wrapped in multiple levels of indirection, making it a somewhat hard to reason about when trying to reconstruct all objects from a given driver.

There's one additional indirection ("namedResources"), right? I don't find it hard to ignore that - but perhaps I am simply used to it. :shrug:

klueska:
I think where it got awkward was when I was tryin to adopt @johnbelamaric*s model to fit into this:
```
type DeviceModel struct {
    metav1.TypeMeta
    metav1.ObjectMeta

    Spec   DeviceModelSpec
    Status DeviceModelStatus
}

type DeviceModelSpec struct {
    NodeName *string
    DriverName string 
    *DeviceModelVariant
}

struct DeviceModelVariant struct {
    DevicePools *DevicePools
    // Other future models
}

type DevicePools struct {
    Pools []DevicePool
}

type DevicePool struct {
...
}
```
These 5 structs turn into one (with multiple instances of the last as top-level objects).

klueska:
I'm not saying we necessarily want to do this, but couldn't the kubelet be updated to whitlelist any new model objects we define (same as the scheduler needing to know to pull them in too)?

johnbelamaric:
> This is what [kubernetes/kubernetes#123699](https://github.com/kubernetes/kubernetes/issues/123699) depends upon: kubelet knows that drivers may access ResourceSlice and ResourceClaim through the REST proxy and denies access to everything else. When combining kubelet from 1.31 with a future DRA driver which wants to publish a "SomeFutureType" instead of a "ResourceSlice", the driver's request will get denied. I'm working on a KEP update for this issue and have a prototype implementation ready that I'll share together with it.
> 
> >

Let's not forget we have versioning in the API. So, we start with v1 without the indirection. One option then is to add a new type; another option is to create DevicePool v2 which in turn has the indirection but roundtrips to v1. Kubelet can use an `Unstructured` for passthrough, so it can still check the types without changes and without versioning issues between apiserver/kubelet/driver. Unless I am missing something.

A couple questions as you think through that design:
- Should kubelet enforce that pool.Driver == the one for the specified driver? Or even just set it, assuming the driver registers with kubelet?
- Similarly, should the pool *name* be something the driver comes up with, or kubelet? Pool names coming from nodes (we could have ones that come from elsewhere) should generally be `<node-name>-< something>`. Are there exceptions to that? Should the driver or the kubelet control it?
-

pohly:
> Kubelet can use an Unstructured for passthrough, so it can still check the types without changes and without versioning issues between apiserver/kubelet/driver. 

kubelet still needs to check whether a request for some unknown resource should be allowed. Let's discuss in https://github.com/kubernetes/enhancements/pull/4615

pohly:
> So, we start with v1 without the indirection. One option then is to add a new type; another option is to create DevicePool v2 which in turn has the indirection but roundtrips to v1.

This doesn't work without the one-of-many concept already being in v1. A new field can be added to v1 to make the round-trip possible, but a client built against v1 without that field will never see it. We have to define in advance how such a client can detect that it doesn't have some information that it should have.

k8srm-prototype/pkg/api/capacity_types_alternate.go
```diff
@@ -0,0 +1,177 @@
...
+	// +optional
+	SharedResourceRequests []DeviceResourceRequest `json:"sharedResourceRequests,omitempty"`
+}
+
+// DeviceResource represents a named resource that is consumable by a device.
+// We indirect through DeviceResourceValue to enforce a one-of-many for the
+// different types of resources this struct can represent.
+type DeviceResource struct {
+	Name string `json:"name"`
+	*DeviceResourceValue `json:",inline"`
```
klueska, line 99: I've noticed that this indirection (and the others like it) still haven't made it into the prototype or been discussed much. I really feel like limiting ourselves to *just* a quantity is going to bite us in the future.

johnbelamaric: I agree, this is probably needed. We'll need to convince @thockin that the complexity is worth it. I think that given that the "capacity" model is primarily for driver developers, a little more complexity is OK. Users will have to be able to understand it though for troubleshooting.

Can you make sure this covered in the "open questions" doc?

thockin: I'm not sure I understand what this is?

PR #9, opened 2024-05-07
Additional PodSpec examples

johnbelamaric:


pohly (commented):
Can you remove out-dated entries from https://github.com/kubernetes-sigs/wg-device-management/blob/main/k8srm-prototype/pod-spec-examples.yaml?

Doesn't have to be in this PR, but it seemed related.

k8srm-prototype/testdata/pod-two-containers-one-gpu.yaml
```diff
@@ -0,0 +1,31 @@
...
+      requests:
+        cpu: 10m
+        memory: 10Mi
+    devices:
+    - name: foozer-gpu
+  deviceClaims:
+  - name: foozer-gpu
+    source:
+      templateClass:
+        className: example.com-foozer
```
pohly, line 31: What is a "template class"? What's the difference compared to a "class"? Which object does this name refer to? A [DeviceClass](https://github.com/kubernetes-sigs/wg-device-management/blob/a42fbe747a5d71fa3cf14bc56ef8170bf0ab036c/k8srm-prototype/pkg/api/claim_types.go#L13C6-L13C17)?

johnbelamaric: It's just one of the forms of ClaimSource: ObjectMeta + ClassName. The idea being that this would be one of the most common things, and in this way we don't need to pre-create a template object.

See https://github.com/kubernetes-sigs/wg-device-management/blob/main/k8srm-prototype/notes-and-open-questions.md#notes

I'll try to write out the ClaimSource object and include a pointer to that.

johnbelamaric: And the discussion here: https://github.com/kubernetes-sigs/wg-device-management/blob/main/k8srm-prototype/testdata/pod-one-container-one-gpu.yaml

johnbelamaric: Ok, I added claim_source.go here in the testdata directory so you can see what I am trying to do.

@klueska this should cover what we discussed yesterday, the idea of some external claim generator (we can add a fieldRef or we can assume a specific field name to drop the `spec` in).

None of this addresses the problem I found with the model, where we need to both do a `matchAttributes` across a set of claims, AND have individually addressable claims to mount into the containers. That likely will need to be done with a second level of "name", so the container mount points become two names not one.

k8srm-prototype/testdata/pod-two-containers-two-gpus-1.yaml
```diff
@@ -0,0 +1,38 @@
...
+  - image: registry.k8s.io/pause:3.6
+    name: my-other-container
+    resources:
+      requests:
+        cpu: 10m
+        memory: 10Mi
+    devices:
+    - name: other-foozer-gpu
+  deviceClaims:
+  ## PROBLEM: We want to specify matchAttributes here but cannot! In the "unrolled"
```
johnbelamaric, line 29: We probably need a second name, inside the claim to reference the "sub claim". Then, the container mount point has to list both names, the main claim and the subclaim.

OR - we only allow one claim in the podspec, and nest all the subclaims in that. Then we only need one claim, it just becomes the entire former "pod spec" piece of this.

k8srm-prototype/testdata/pod-two-containers-one-gpu.yaml
johnbelamaric: Due to the versioning concerns in pod spec, I am wondering if having a single reference in pod spec to a claim object that contains all the various options is better. This way, the container spec still just uses one name. Later, if we can move everything to the pod spec, evolve the device claim to be more sensible, either in a beta version or even in a v2.

 It's frustrating to be locked in by the versioning scheme :( - I hate that our permanent UX is affected by temporary considerations.

PR #10, opened 2024-05-07
Fix go.mod

johnbelamaric:
Not sure how this didn't get pushed before.


johnbelamaric:
@klueska @pohly FYI somehow this was working locally for me but this change never made it in. I presume "make" wasn't working for you?

PR #13, opened 2024-05-09
Suggested admin access note

johnbelamaric:
@klueska please lgtm if this matches what you and I discussed

### DIFF
--- a/k8srm-prototype/open-questions.md
+++ b/k8srm-prototype/open-questions.md
@@ -3,6 +3,18 @@
 - Will the inlined pointer be OK for the DeviceClaimInstance one-of?
 - The way management pods are handled is clunky at best. A separate top-level
   type still seems cleaner.
+  - klueska: Another way to handle this, based on how it is done today in
+    classic DRA, would be the following:
+    - Move the AdminAccess flag back into the class.
+    - We define a particular annotation, something like `device-class-taint` on
+      the class.
+    - When a user uses a tainted class in a claim, a corresponding namespace
+      annotation, something like `device-class-toleration` must be present on
+      the namespace, or the scheduler will not satisfy the claim.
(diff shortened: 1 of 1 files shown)


k8srm-prototype/open-questions.md
```diff
@@ -3,6 +3,18 @@
 - Will the inlined pointer be OK for the DeviceClaimInstance one-of?
 - The way management pods are handled is clunky at best. A separate top-level
   type still seems cleaner.
+  - klueska: Another way to handle this, based on how it is done today in
+    classic DRA, would be the following:
+    - Move the AdminAccess flag back into the class.
+    - We define a particular annotation, something like `device-class-taint` on
+      the class.
+    - When a user uses a tainted class in a claim, a corresponding namespace
+      annotation, something like `device-class-toleration` must be present on
```
klueska, line 12: I pictured it more as the class "defining" the name of the annotation that the namespace must have (e.g. `gpu.nvidia.com-management-access`) and then the namespace must have that annotation to be able to instantiate clams from that class.

So like:
```
AccessMode: Monitoring
NamespaceAnnotationRequiredForUse: gpu.nvidia.com-management-access
```

Another related thing is that we typically will want to monitor either (1) all devices, or (2) the devices granted to a specific claim. We probably don't want (or need) to do selection on devices ourselves for these kind of claims. But I think we will want the ability to point at an already allocated claim and say that we are monitoring the devices granted to that. This can likely wait, so long as we see a path for including it in the future.

I see this specifically being useful for running the MPS server required to manage the set of GPUs allocated with MPS sharing enabled. At present, I give each MPS server access to all GPUs and have to do some hacky things inside the container to limit its view of the GPUs it is supposed to govern. It would be great if we had integrated support to limit this view.

Issue #21, opened 2024-06-05
dra-evolution: network-attached devices


pohly:
This was excluded from https://github.com/kubernetes-sigs/wg-device-management/pull/14 to limit the scope. It's probably not going to be included in in 1.31 because of time constraints and because it's lower priority. Classic DRA can handle network-attached devices, if we [restore support](https://github.com/kubernetes-sigs/wg-device-management/issues/17) for it.

/assign

pohly:
Basic support for network-attached devices got included in 1.31 when we designed the APIs to allow node selectors.

/close

PR #14, opened 2024-05-14
DRA API evolution

pohly:
This creates a "dra-evolution" directory with an alternative to the "k8srm-prototype". Whereas"k8srm-prototype" started with a blank slate, "dra-evolution" started with the 1.30 API and got updated to make it more like "k8srm-prototype".

It includes respectively supersedes the following incremental KEP updates:
- https://github.com/kubernetes/enhancements/pull/4611
- https://github.com/kubernetes/enhancements/pull/4613
- https://github.com/kubernetes/enhancements/pull/4614
- https://github.com/kubernetes/enhancements/pull/4619
- https://github.com/kubernetes/enhancements/pull/4612

The top-level README.md has a comparison table with a rationale why certain changes were made (or not made).

It's now possible to do `go test ./testdata`, which checks that the YAML files and the embedded CEL expressions match the types and CEL syntax.

`cmd` and `pkg/gen` have not been updated yet. The `pkg/api` is complete and now also includes the pod spec changes. *But please don't review yet*. @pohly and @klueska are still in the process of converting the YAML testdata files, which may lead to changes in the API.

PR #26, opened 2024-06-06
dra-evolution: restore minimal support for "classic DRA"

pohly:
"Classic DRA" has a control plane controller which handles allocation and deallocation of a claim in cooperation with the scheduler. That is the part which gets added back to the API.

What doesn't get restored compared to v1.30.0 is:
- immediate allocation
- using a CRD as user-facing API
- the opaque "resource handle" in the allocation result

Drivers which want to support this mode of operation have to use the in-tree types. The "config" fields support that, albeit perhaps with a slightly less natural API.

While at it, some occurrences of "resource" get replaced with "device" or "DRA".

/assign @thockin @klueska

### DIFF
--- a/dra-evolution/pkg/api/claim_types.go
+++ b/dra-evolution/pkg/api/claim_types.go
@@ -5,6 +5,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/validation"
 )
 
 // DeviceClass is a vendor or admin-provided resource that contains
@@ -401,14 +402,27 @@ type RequestAllocationResult struct {
 	// device to be allocated.
 	RequestName string `json:"requestName"`
 
-	// This node name together with the driver name and
-	// the device name field identify which device was allocated.
-	NodeName string `json:"nodeName"`
+	// This name together with the driver name and the device name field
+	// identify which device was allocated.
+	//
+	// For a driver with node-local devices, it is the name of the node and
+	// the overall ID is `<driver name>/<node name>/<device name>`.
+	//
+	// For other drivers, the overall ID is `<driver name>/<device pool
+	// name>/<device name>`.
+	//
+	// Must not be longer than 253 characters and may contain one or more
+	// DNS sub-domains separated by slashes.
+	//
+	// +optional
+	DevicePoolName string `json:"devicePoolName,omitempty"`
 
 	// DeviceName references one device instance via its name in the driver's
-	// resource pool.
+	// resource pool. It must be a DNS label.
 	DeviceName string `json:"deviceName"`
 }
+
+const DevicePoolNameMaxLength = validation.DNS1123SubdomainMaxLength // Same as for a single node name.
 
 // DeviceConfiguration is one entry in a list of configuration pieces for a device.
 type DeviceConfiguration struct {
--- a/dra-evolution/pkg/api/claim_types.go
+++ b/dra-evolution/pkg/api/claim_types.go
@@ -322,8 +322,8 @@ type AllocationResult struct {
 	// involving a driver.
 	//
 	// A driver may allocate devices provided by other drivers, so this
-	// driver name here can be different from the driver names listed in
-	// DriverData.
+	// driver name here can be different from the driver names listed for
+	// the results.
 	//
 	// This is an alpha field and requires enabling the DRAControlPlaneController
 	// feature gate.
@@ -353,27 +353,6 @@ type AllocationResult struct {
 	//
 	// +optional
 	Shareable bool `json:"shareable,omitempty" protobuf:"varint,3,opt,name=shareable"`
-}
-
-// DriverData holds information for processing by a specific kubelet plugin.
-type DriverData struct {
-	// DriverName specifies the name of the DRA driver whose kubelet
-	// plugin should be invoked to process the allocation once the claim is
-	// needed on a node.
-	//
-	// Must be a DNS subdomain and should end with a DNS domain owned by the
-	// vendor of the driver.
-	DriverName string `json:"driverName" protobuf:"bytes,1,name=driverName"`
-
-	// Config contains all the configuration pieces that apply to the entire claim
-	// and that were meant for the driver which handles these devices.
-	// They get collected during the allocation and stored here
-	// to ensure that they remain available while the claim is allocated.
-	//
-	// Entries are listed in the same order as in claim.config.
-	//
-	// +optional
-	Config []DriverConfigurationParameters `json:"config,omitempty"`
 }
 
 // RequestAllocationResult contains configuration and the allocation result for
--- a/dra-evolution/pkg/api/claim_types.go
+++ b/dra-evolution/pkg/api/claim_types.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -33,6 +34,21 @@ type DeviceClass struct {
 	// +optional
 	// +listType=atomic
 	Requirements []Requirement `json:"requirements,omitempty" protobuf:"bytes,4,opt,name=requirements"`
+
+	// Only nodes matching the selector will be considered by the scheduler
+	// when trying to find a Node that fits a Pod when that Pod uses
+	// a claim that has not been allocated yet *and* that claim
+	// gets allocated through a control plane controller. It is ignored
+	// when the claim does not use a control plane controller
+	// for allocation.
+	//
+	// Setting this field is optional. If unset, all nodes are candidates.
+	//
+	// This is an alpha field and requires enabling the DRAControlPlaneController
+	// feature gate.
+	//
+	// +optional
+	SuitableNodes *v1.NodeSelector `json:"suitableNodes,omitempty" protobuf:"bytes,5,opt,name=suitableNodes"`
 }
 
 // ConfigurationParameters must have one and only one field set.
@@ -162,6 +178,19 @@ type ResourceClaimSpec struct {
 	// by multiple consumers at the same time.
 	// +optional
 	Shareable bool `json:"shareable,omitempty" protobuf:"bytes,3,opt,name=shareable"`
+
+	// ControllerName defines the name of the DRA driver that is meant
+	// to handle allocation of this claim. If empty, allocation is handled
+	// by the scheduler while scheduling a pod.
+	//
+	// Must be a DNS subdomain and should end with a DNS domain owned by the
+	// vendor of the driver.
+	//
+	// This is an alpha field and requires enabling the DRAControlPlaneController
+	// feature gate.
+	//
+	// +optional
+	ControllerName string `json:"controllerName,omitempty"`
 }
 
 // Constraint must have one and only one field set.
@@ -187,8 +216,8 @@ type Constraint struct {
 }
 
 // Request is a request for one of many resources required for a claim.
-// This is typically a request for a single resource like a device, but can
-// also ask for one of several different alternatives.
+// This is typically a request for a single resource like a device, but may get
+// extended to also ask for one of several different alternatives.
 type Request struct {
 	// The name can be used to reference this request in a pod.spec.containers[].resources.claims
 	// entry.
@@ -249,7 +278,7 @@ type ResourceRequestDetail struct {
 	Requirements []Requirement `json:"requirements,omitempty" protobuf:"bytes,4,opt,name=requirements"`
 }
 
-// ResourceClaimStatus tracks whether the resource has been allocated and what
+// ResourceClaimStatus tracks whether the claim has been allocated and what
 // the result of that was.
 type ResourceClaimStatus struct {
 	// Allocation is set once the claim has been allocated successfully.
@@ -269,10 +298,38 @@ type ResourceClaimStatus struct {
 	// +patchMergeKey=uid
 	// +optional
 	ReservedFor []ResourceClaimConsumerReference `json:"reservedFor,omitempty" protobuf:"bytes,3,opt,name=reservedFor" patchStrategy:"merge" patchMergeKey:"uid"`
+
+	// Indicates that a claim is to be deallocated. While this is set,
+	// no new consumers may be added to ReservedFor.
+	//
+	// This is only used if the claim needs to be deallocated by a DRA driver.
+	// That driver then must deallocate this claim and reset the field
+	// together with clearing the Allocation field.
+	//
+	// This is an alpha field and requires enabling the DRAControlPlaneController
+	// feature gate.
+	//
+	// +optional
+	DeallocationRequested bool `json:"deallocationRequested,omitempty" protobuf:"varint,4,opt,name=deallocationRequested"`
 }
 
 // AllocationResult contains attributes of an allocated resource.
 type AllocationResult struct {
+	// ControllerName is the name of the DRA driver which handled the
+	// allocation. That driver is also responsible for deallocating the
+	// claim. It is empty when the claim can be deallocated without
+	// involving a driver.
+	//
+	// A driver may allocate devices provided by other drivers, so this
+	// driver name here can be different from the driver names listed in
+	// DriverData.
+	//
+	// This is an alpha field and requires enabling the DRAControlPlaneController
+	// feature gate.
+	//
+	// +optional
+	ControllerName string `json:"controllerName,omitempty"`
+
 	// DriverData contains the state associated with an allocation that
 	// should be maintained throughout the lifetime of a claim. Each
 	// entry contains data that should be passed to a specific kubelet
@@ -286,20 +343,20 @@ type AllocationResult struct {
 	// +optional
 	DriverData []DriverData `json:"driverData,omitempty" protobuf:"bytes,1,opt,name=driverData"`
 
-	// This is the node which provides the allocated devices.
+	// Setting this field is optional. If unset, the allocated devices are available everywhere.
 	//
-	// If empty, then the devices are not local to one particular node.
-	NodeName string `json:"nodeName,omitempty" protobuf:"bytes,2,opt,name=nodeName"`
+	// +optional
+	AvailableOnNodes *v1.NodeSelector `json:"availableOnNodes,omitempty" protobuf:"bytes,2,opt,name=availableOnNodes"`
 
-	// Shareable determines whether the resource supports more
-	// than one consumer at a time.
+	// Shareable determines whether the claim supports more than one consumer at a time.
+	//
 	// +optional
 	Shareable bool `json:"shareable,omitempty" protobuf:"varint,3,opt,name=shareable"`
 }
 
 // DriverData holds information for processing by a specific kubelet plugin.
 type DriverData struct {
-	// DriverName specifies the name of the resource driver whose kubelet
+	// DriverName specifies the name of the DRA driver whose kubelet
 	// plugin should be invoked to process the allocation once the claim is
 	// needed on a node.
 	//
@@ -307,20 +364,8 @@ type DriverData struct {
 	// vendor of the driver.
 	DriverName string `json:"driverName" protobuf:"bytes,1,name=driverName"`
 
-	// Data contains all information about the allocation that the kubelet
-	// plugin will need.
-	//
-	// +optional
-	Data *StructuredDriverData `json:"data,omitempty" protobuf:"bytes,2,opt,name=data"`
-
-	// Alternative to "Data", not describe further here for the sake of brevity:
-	// OpaqueData *string // Set by control plane controller when using "classic DRA".
-}
-
-// StructuredDriverData is the in-tree representation of the allocation result.
-type StructuredDriverData struct {
 	// Config contains all the configuration pieces that apply to the entire claim
-	// and that were meant for the driver which handles these resources.
+	// and that were meant for the driver which handles these devices.
 	// They get collected during the allocation and stored here
 	// to ensure that they remain available while the claim is allocated.
 	//
@@ -329,7 +374,7 @@ type StructuredDriverData struct {
 	// +optional
 	Config []DriverConfigurationParameters `json:"config,omitempty"`
 
-	// Results lists all allocated driver resources.
+	// Results lists all allocated devices.
 	//
 	// +listType=atomic
 	Results []RequestAllocationResult `json:"results" protobuf:"bytes,4,name=results"`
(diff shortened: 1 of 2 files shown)


dra-evolution/pkg/api/claim_types.go
```diff
@@ -286,41 +343,29 @@ type AllocationResult struct {
 	// +optional
 	DriverData []DriverData `json:"driverData,omitempty" protobuf:"bytes,1,opt,name=driverData"`
 
-	// This is the node which provides the allocated devices.
+	// Setting this field is optional. If unset, the allocated devices are available everywhere.
 	//
-	// If empty, then the devices are not local to one particular node.
-	NodeName string `json:"nodeName,omitempty" protobuf:"bytes,2,opt,name=nodeName"`
+	// +optional
+	AvailableOnNodes *v1.NodeSelector `json:"availableOnNodes,omitempty" protobuf:"bytes,2,opt,name=availableOnNodes"`
```
pohly, line 350: This is the only place where the "structured parameters" API gets changed. I am bringing back the node selector and moved the "node name" part down into `RequestAllocationResult` where it sits alongside the device name.

A simple node name is not enough because "classic DRA" is how network-attached devices can be supported in 1.31.

dra-evolution/pkg/api/claim_types.go
```diff
@@ -286,41 +343,29 @@ type AllocationResult struct {
...
 	DriverName string `json:"driverName" protobuf:"bytes,1,name=driverName"`
 
-	// Data contains all information about the allocation that the kubelet
-	// plugin will need.
-	//
-	// +optional
-	Data *StructuredDriverData `json:"data,omitempty" protobuf:"bytes,2,opt,name=data"`
-
-	// Alternative to "Data", not describe further here for the sake of brevity:
-	// OpaqueData *string // Set by control plane controller when using "classic DRA".
```
pohly, line 317: I came to the conclusion that we don't need this opaque data and thus `StructuredDriverData` can be inlined.

A control plane controller can and should use instead the in-tree RequestAllocationResult. Then selecting a specific request for a container works the same way in both allocation modes.

dra-evolution/pkg/api/claim_types.go
pohly: In the first commit I kept the status structured as it was before. But we can go further and do what @johnbelamaric suggested and make the status more like the spec, with a list of results at the root instead of nesting those under each driver.

The second commit does that. It's not directly related to classic DRA, it just was motivated by the previous change because that one already put "nodeName" and "deviceName" into the same struct. Now that struct also has "driverName", which completes the triple that is needed to uniquely identify the device.

dra-evolution/pkg/api/claim_types.go
```diff
@@ -361,9 +401,12 @@ type RequestAllocationResult struct {
 	// device to be allocated.
 	RequestName string `json:"requestName"`
 
-	// DeviceName references one device instance via its name in the resource
-	// pool. Driver name and node name have to be determined from the
-	// context.
+	// This node name together with the driver name and
+	// the device name field identify which device was allocated.
+	NodeName string `json:"nodeName"`
```
pohly: Let's generalize this a bit more so that it also works for network-attached devices, whether they get allocated through a control plane controller (this would be possible in 1.31) or with some future extension of structured parameters.

```suggestion
	// This name together with the driver name and
	// the device name field identify which device was allocated.
	//
	// For a driver with node-local devices, it is the name of the node
	// and the overall ID is `<driver name>/<segment name = node name>/<device name>`.
	//
	// For a driver where device names are unique in a certain subset of
	// the cluster (region, rack, etc.), the name identifies that segment.
	//
	// For a driver where device names are unique in the cluster, this
	// name may be empty.
	// 
	// +optional
	SegmentName string `json:"segmentName,omitempty"`
```

dra-evolution/testdata/pod-one-container-one-gpu-one-vf.yaml
```diff
@@ -63,17 +63,23 @@ spec:
     - deviceSelector: "device.stringAttributes['sriov-nic.example.org/sriovType'] == 'vf'"
 status:
   allocation:
-    nodeName: worker-1
+    availableOnNodes:
```
johnbelamaric, line 68: I'm confused. Why is this here, if this is not an example of a classic DRA allocation? It should not be needed except for classic DRA, right?

dra-evolution/pkg/api/claim_types.go
```diff
@@ -269,85 +298,96 @@ type ResourceClaimStatus struct {
...
-	// plugin once the claim lands on a node.
+	// ControllerName is the name of the DRA driver which handled the
+	// allocation. That driver is also responsible for deallocating the
+	// claim. It is empty when the claim can be deallocated without
+	// involving a driver.
 	//
-	// Setting this field is optional. It has a maximum size of 32 entries.
-	// If empty, nothing was allocated for the claim and kubelet does not
-	// need to prepare anything for it.
+	// A driver may allocate devices provided by other drivers, so this
```
johnbelamaric, line 324: That sounds tricky. The two drivers then somehow need to coordinate on the allocations? I don't think we need to get deep into this now, but I'd like to understand what this use case is so if necessary we can eventually fold it into the structured parameters model.

dra-evolution/pkg/api/claim_types.go
johnbelamaric: I still find this confusing. I would think the default case would be "available only on the node named in the result" rather than "available everywhere". This means we have to set this for every single result, even though none in the beta will actually need it?

Actually, I am not sure this is correct even - shouldn't this be on a per request basis? Or I am not understanding something. For example, couldn't a single claim contain requests for two different network attached devices, which are available on different sets of nodes? In that case, wouldn't we need this in the request result?

I also am not clear on why we need this in the *result* at all. The result is used to tell the driver how to actuate the devices. This field here suggests that somehow a subsequent request for a device may be resolved (by the platform, not the user) to this claim? I think that's the key thing I am missing. This is used for some use case where a controller manages access to a given set of devices that are pre-allocated by a claim? If that's the case, the ReservedFor maybe makes more sense now too.

I am not sure that stuff belongs in claim.

I will say again - I think the network attached stuff is really really different, I am not sure a single API for both node-local and network attached is useful.

dra-evolution/pkg/api/claim_types.go
pohly: The "meta driver" which implements the allocation needs to use the same allocation API as the drivers that it is wrapping. @klueska used this to combine GPUs with NICs. In practice, the "allocation API" was the CRD that those two drivers used to advertise their node-local devices.

I think we have that use case covered in some form with match attributes. The "reserve a device in ResourcePool" could become our standardized "allocation API" for cases where drivers want to support both structured parameters for simple cases and a control plane controller for the more complex scenarios. As it stands now, it's either/or.

dra-evolution/testdata/pod-one-container-one-gpu-one-vf.yaml
pohly: I am preparing `resourceClaim.status` for network-attached devices here. We didn't have those at all before. Now we get support for them through classic DRA and eventually through structured parameters when we get to https://github.com/kubernetes-sigs/wg-device-management/issues/21.

pohly: I simplified this field in https://github.com/kubernetes-sigs/wg-device-management/pull/14 because I wanted to avoid the distraction (minimal API first, extensions later). This is one of those extensions.

dra-evolution/pkg/api/claim_types.go
pohly: I'm torn between allowing just one intermediate segment and many. It could make sense to allow `<driver name>/<region>/<data center>/<device name>` as the string ID for a device.

We want to be sure that there is a driver name and device name, but what the driver puts into the middle doesn't really matter.

Naming, oh why do you have to be so difficult?!

Using "PoolName" is tempting ("names the pool in which the device name must be unique"), but it clashes with `ResourcePool.Name`.

`DevicePoolName`? Max length 128 characters, DNS sub-domains separated by slashes. Optional.

That goes well together with `DeviceName`.

Once we get to structured parameters for network-attached devices, a `ResourcePool` object for those would have:
- a `NodeSelector *v1.NodeSelector instead of `NodeName string`.
- such a `DevicePoolName` that serves as part of the unique ID

dra-evolution/pkg/api/claim_types.go
pohly: > This means we have to set this for every single result, even though none in the beta will actually need it?

The beta uses this to determine which devices are allocated. For each result, it needs to connect the result back to the corresponding ResourcePool device. We do this via the `<device name>/<node name>/<device name>` triple, so that has to be stored somewhere.

That we now have `<node name>` repeated for each entry is a side effect of flattening the status. It is a bit redundant if all that the claim uses are node-local devices from the same node, but that isn't necessarily always the case. Let me rename the result field as suggested in https://github.com/kubernetes-sigs/wg-device-management/pull/26#discussion_r1630785781, then it might look better.

> I think the network attached stuff is really really different, I am not sure a single API for both node-local and network attached is useful

I see it the other way around. Node-local is a special case and with a little of "out of the box" (sic) thinking we can generalize the API so that it works for both.

dra-evolution/pkg/api/claim_types.go
pohly: We have to allow a node name, so the max length has to be 253.

pohly: I've pushed this as a third commit.

dra-evolution/testdata/pod-one-container-one-gpu-one-vf.yaml
```diff
@@ -57,23 +57,31 @@ spec:
   constraints:
   - matchAttribute: dra.k8s.io/pcie_root
   requests:
-  - deviceClassName: foozer.example.com
-  - deviceClassName: sriov-nic-example.org
+  - name: gpu
+    deviceClassName: foozer.example.com
```
pohly, line 61: We decided to make request names mandatory. Not all examples have been updated, so I am doing it here now because I am touching it.

dra-evolution/pkg/api/claim_types.go
pohly: And here's how this would work together with a ResourcePool (but note [my question about that name](https://github.com/kubernetes-sigs/wg-device-management/pull/28/files#r1630964004):

https://github.com/kubernetes-sigs/wg-device-management/blob/78e7edd841422bd887e61061778213ad1e2ea949/dra-evolution/pkg/api/capacity_types.go#L9-L21

https://github.com/kubernetes-sigs/wg-device-management/blob/78e7edd841422bd887e61061778213ad1e2ea949/dra-evolution/pkg/api/capacity_types.go#L50-L102

dra-evolution/pkg/api/claim_types.go
johnbelamaric: I think I am still confused as to what this node selector is for. There are two possibilities (at least), and I don't understand which is the intent:

1. Allowing us to determine the resource pool from which this result is allocated. For node-local devices, this is not needed because NodeName is in each entry. For networked devices, a node selector seems insufficiently precise; instead I would expect some identifier for the networked pool itself (if we want to allow rebalancing, it can't be the k8s name of the pool...but do we need to allow that?).
2. Given the thinking in 1), that node selector is insufficiently precise, I thought maybe you intended this to be used to "advertise" the availability of a pre-allocated claim. That's what I was trying to understand in the last comment.


It sounds, though, like maybe it is 1) after all. But I don't see how a node selector is the right thing to use as an identifier for pool of networked devices. I don't think a node selector is a good identifier.

Or am I completely misunderstanding?

pohly: It is not 1. The purpose of the field is to tell a scheduler which sees an already allocated ResourceClaim where the devices are available. This is relevant for sharing an allocated claim between pods. Typically this is a node, but it doesn't have to be.

What identifies the device and thus the pool is the `driver name/device pool name/device name` combination, which is stored for each result. Driver and pool can be different for each result once we allow network-attached devices.

pohly: One could argue that by looking at all devices, a scheduler can infer where they are available. But that only works when the ResourcePool currently exists, which is not required and cannot be enforced after allocation. With classic DRA, the scheduler cannot look up further information about devices and has to be given this node selector.

IMHO, an explicit field is also easier for users.

dra-evolution/pkg/api/claim_types.go
johnbelamaric: I think with the partitionable device model, we can handle this completely on the driver side. See https://github.com/kubernetes-sigs/wg-device-management/pull/27#issuecomment-2155167940. Is this still needed in that case?

pohly: We need the controller name for classic DRA. It cannot be inferred from something else.

dra-evolution/pkg/api/claim_types.go
pohly: Let me explain this differently. A claim status must answer two questions:
- Which nodes have access to the devices?
- Which devices were allocated?

In the special case of node-local devices, the answer to both was a node name. In the more general case, we need a node selector for the first question and some unique ID for a device. https://github.com/kubernetes-sigs/wg-device-management/pull/26#discussion_r1630073547 is about that.

klueska: The way I've always thought about it is this:
1. A vendor-specific controller is the one in charge of doing the allocation of this claim
2. It may choose to make the resource available on more than one node (if possible)
3. It sets the node selector to reflect the set of nodes where the resource can/will be made available
4. The scheduler then has the freedom to schedule a pod referencing this claim on any of the nodes in the node selector (rather than just one specific node)

dra-evolution/pkg/api/claim_types.go
klueska: The point here is that there are two different ResourcePools -- each managed by a "standard" driver for devices of that type. These drivers can be from different vendors and don't have to be aware of each other.

Along side of this, one can then create a "meta" controller which knows how to allocate devices of both types together. So long as it knows how to speak the "language" of each driver in order to perform these allocations atomically, it can sit side by side with them without interfering with the operation of each "standard" driver.

We do not envision such "meta" controllers being for general purpose use, but rather customized to specific environments that need to do non-standard things. This mechanims allows them to install the standard drivers on their clusters, and then only deploy a custom controller to do allocations, rather than having to write an entirely new driver from scratch that also implements the kubelet-plugin side of things.

dra-evolution/pkg/api/claim_types.go
klueska: So if I'm reading this diff correctly, the only fields in this struct now are `DriverName` and `Config`? Is this because the new `Config` section is (under the hood) just a RawExtension, so that can now be used to encapsulate any opaque data that needs to be passed through?

dra-evolution/pkg/api/claim_types.go
```diff
@@ -269,85 +299,96 @@ type ResourceClaimStatus struct {
...
 // RequestAllocationResult contains configuration and the allocation result for
 // one request.
 type RequestAllocationResult struct {
+	// DriverName specifies the name of the DRA driver whose kubelet
+	// plugin should be invoked to process the allocation once the claim is
+	// needed on a node.
+	//
+	// Must be a DNS subdomain and should end with a DNS domain owned by the
+	// vendor of the driver.
+	DriverName string `json:"driverName" protobuf:"bytes,1,name=driverName"`
```
klueska, line 367: Why do we need this both here and in `DriverData`? Speaking of which, I don't see `DriverData` embedded in any of the other types here. Remind me where it gets referenced again?

dra-evolution/pkg/api/claim_types.go
```diff
@@ -361,12 +402,28 @@ type RequestAllocationResult struct {
...
+	// the overall ID is `<driver name>/<node name>/<device name>`.
+	//
+	// For other drivers, the overall ID is `<driver name>/<device pool
+	// name>/<device name>`.
+	//
+	// Must not be longer than 253 characters and may contain one or more
+	// DNS sub-domains separated by slashes.
+	//
+	// +optional
+	DevicePoolName string `json:"devicePoolName,omitempty"`
```
klueska, line 397: When I see this I think of out whole `ResourcePool` and `DevicePool` discussion and it makes me think this is referring to the name of the `ResourcePool` that the device in this result was allocated from. I think it is probably meant to be something different though, but the previous discussions are carrying too much baggage for me to understand exactly what that is.

dra-evolution/pkg/api/claim_types.go
```diff
@@ -376,6 +433,14 @@ type DeviceConfiguration struct {
...
 }
 
+// DriverConfigurationParameters must have one and only one one field set.
+//
+// In contrast to ConfigurationParameters, the driver name is
+// not included and has to be infered from the context.
+type DriverConfigurationParameters struct {
+	Opaque *runtime.RawExtension `json:"opaque,omitempty" protobuf:"bytes,1,opt,name=opaque"`
+}
+
```
klueska, line 422: I don't quite understand this comment. The `ConfigurationParameters` and `DriverConfigurationParameters` structs look identical to me.

klueska: Edit: I see the difference now -- the other struct has an embedded `OpaqueConfigurationParameters` with an embedded `DriveName` field. rather than a direct field for `runtime.RawExtension`.

It's still confusing to look at, but I at least see the difference now...

dra-evolution/pkg/api/claim_types.go
johnbelamaric: Thanks, I do wonder if the alternative "compound device" design above will cover that use case. Nonetheless, we can keep this here for all classic DRA use cases and we will see if we need it in the long run.

dra-evolution/pkg/api/claim_types.go
johnbelamaric: Thank you, those explanations help a lot. This lgtm.

dra-evolution/pkg/api/claim_types.go
johnbelamaric: I find it super confusion to have both ResourcePool and DevicePool. I think we can find a better name.

Based on the discussion in the other comment, the purpose of this is to allow the controller to identify the "owner" of the device. In the case of node-local, that is always the node so we didn't need anything but node name. With networked devices, this is controller-specific and effectively an opaque (to K8s) identifier. I can see where you are going with "device pool" in that sense, but I think it's too confusing. So, let's brainstorm some other ideas. What about something of this form: `{"",Device,Resource} + {Manager,Owner,Driver} + {Id,Identifier,Key,Token,OpaqueId}`. So, like, `DriverOpaqueId` would be one possible name based on that. Other ideas?

johnbelamaric: Actually I really like `DriverOpaqueId`. It makes it clear it is 1) driver-specific; 2) opaque to k8s (not a resource name for example); 3) an identifier

pohly: I agree, the naming of ResourcePool needs further thought. The current name implies "this is one pool" but it's really just a subset of some other, larger pool, the one which contains all devices provided by the same driver name and published with the same "DriverOpaqueID" (or whatever we call it).

I would stay with the ResourceSlice type and define it like this:
> Devices are identified by the name of the driver, a pool name, and a unique name inside that pool. A driver publishes information about its resources in one or more ResourceSlice objects. Each slice is part of one pool and contains the common attributes of a pool (the pool name, which nodes have access to the devices in the pool) and some of the devices in the pool. The common attributes are provided in each slice to enable correlating them (via the driver name and pool name) and to provide all relevant information without having to look up other objects.
>
> Another common attribute is the total number of devices in the pool. This information enables consumers to determine whether they have seen all slices of a pool.
>
> A slices belonging to the same pool must share the same generation number (TODO: can we use ObjectMeta.Generation for this or do we have to define our own field?). When a consumer sees slices of different generations, it should ignore all old slices. This can happen while the driver is updating the pool.
>
> Pools that are local to a node are special. They get published with a node name, which then also serves as pool name. There are special permission checks that limit access from a node to slices with the same node name. Provisioning a new node of the same type is likely to provide the same set of devices as some old node.


With this definition, we can safely implement "give me access to all devices" because scheduler knows when it lacks information. The "pool name" is meaningful and won't get confused with the name of a ResourcePool.

This begs the question whether we should have a separate "pool" type. I don't think we need it, some small amount of duplicated information IMHO is fine.

dra-evolution/pkg/api/claim_types.go
pohly: Yes, that was the idea.

dra-evolution/pkg/api/claim_types.go
johnbelamaric: Ahh...yes, this makes sense to me. What if we took this a little further, and we *do* have two types: DevicePool and ResourceSlice. Instead of "Pools that are local to a node are special. They get published with a node name, which then also serves as pool name.", we could have the idea of a "DriverInstance" which could be the node name for node-local devices, but could vary with other types of devices. So, it would be the four values: `(DriverName, DriverInstance, DevicePoolName, DeviceName)` that identifies an individual device. As you say, `ResourceSlice` would be the mechanism by which we publish the *elements* of the DevicePool. We could publish a DevicePool resource as well, and it would include a total device count and/or a slice count, and a generation ID as you say, so that we have this one object that we can use to ensure consistency across all the slices for that pool. We could also normalize common attributes up to the pool, as we have discussed before; I don't think we would need to normalize them up to the *slice*, just to the *pool*.

So - I think this is a good model. What I really like is it weakens the "node" boundary in a way I think will be important in the long run (it's even in our charter...). What I *don't* like, is that it requires a lot of explanation. I think if we omit the DevicePool object itself, it will be harder for people to understand. So I would want to have that in there. I think it also helps with revision control - it lets us know what the current generation is. Otherwise, I imagine some race conditions when we just say it is "the latest one we have seen".

dra-evolution/pkg/api/claim_types.go
pohly: `DriverName` became obsolete because I moved all of its fields directly into `AllocationResult`. I've removed it now.

dra-evolution/pkg/api/claim_types.go
pohly: https://github.com/kubernetes-sigs/wg-device-management/pull/26#discussion_r1632105483

dra-evolution/pkg/api/claim_types.go
pohly: Would it make more sense to repeat the driver name, i.e. use `ConfigurationParameters`?

That could look odd in YAML because the driver name will be more obviously redundant:
```yaml
results:
- driverName: dra.example.com
  config:
  - opaque:
      driverName: dra.example.com
      parameters:
        ...
```

pohly: Perhaps it actually makes sense? One driver could understand the parameters of another.

We don't do that with structured parameters (it's defined as "only configs for the driver are passed down to it"), but perhaps that could change and a control plane controller could already do it now.

dra-evolution/pkg/api/claim_types.go
pohly: I've added these ideas with a different wording to https://github.com/kubernetes/enhancements/pull/4709.

Let's discuss there.

dra-evolution/pkg/api/claim_types.go
pohly: Not added yet to https://github.com/kubernetes/enhancements/pull/4709 but perhaps we should?

If we do, we probably should make it consistent and also pass down all configurations down when using structured parameters.

PR #28, opened 2024-06-07
WIP: dra-evolution: network-attached devices

pohly:
This builds on top of the support for network-attached devices in classic DRA (https://github.com/kubernetes-sigs/wg-device-management/pull/26).

I'm not proposing this for 1.31. I just want to show what the ResourcePool change could look like.

### DIFF
--- a/dra-evolution/pkg/api/capacity_types.go
+++ b/dra-evolution/pkg/api/capacity_types.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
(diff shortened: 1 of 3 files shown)


dra-evolution/pkg/api/capacity_types.go
```diff
@@ -46,8 +51,14 @@ type ResourcePoolSpec struct {
...
-	// This is currently required, but this might get relaxed in the future.
-	NodeName string `json:"nodeName"`
+	// Node name and device pool are mutually exclusive. At least one must be set.
+	NodeName string `json:"nodeName, omitempty"`
+
+	// DevicePool can be used for non-instead of a NodeName to define where the devices
+	// are available.
+	//
+	// Node name and device pool are mutually exclusive. At least one must be set.
+	DevicePool *DevicePool `json:"devicePool,omitempty"`
```
pohly, line 61: I know I accepted the "ResourceSlice" -> "ResourcePool" renaming, but when I spell it out like this, I wonder again whether "ResourceSlice" isn't the better name after all.

"Slice" here is not a Go slice, it's a subset of some larger pool which provides several different devices. *That* pool is the device pool from which devices get allocated, not the individual object.

In https://github.com/kubernetes-sigs/wg-device-management/pull/27, @johnbelamaric said "each partitionable device needs to be its own pool". If that is how it plays out, then naming the API type "ResourcePool" makes less sense to me.

/cc @klueska @johnbelamaric

pohly:
/close

Enough got into Kubernetes already for this.

Issue #20, opened 2024-06-05
dra-evolution: partitioning of devices


pohly:
This was excluded from https://github.com/kubernetes-sigs/wg-device-management/pull/14 to limit the scope. It's a stretch goal for 1.31.

/assign @klueska @johnbelamaric

johnbelamaric:
In the 1.31 KEP, we included the APIs defined in https://github.com/kubernetes-sigs/wg-device-management/pull/27, but Mrunal and Tim [raised legitimate concerns](https://github.com/kubernetes/enhancements/pull/4709#pullrequestreview-2114492231) with the ...verbosity... of that API.

I can think of a few alternatives we can debate, and will propose them as separate PRs in this repo.

cc @thockin @mrunalp @pohly @klueska

johnbelamaric:
Here are some options. Each of the options 2+ are built *on top* of option 1.

I suggest looking at the file `dra-evolution/testdata/pools-two-nodes-dgxa100.yaml` in each PR. This is an example YAML for two 8 GPU servers based on the NVIDIA simulated devices. Real world will be similar, but probably add MORE attributes.

**Option 1**
* As in the 1.31 KEP
* No attributes shared among devices
* Flattens (de-normalizes) all physical GPU shared resources into a single array of SharedCapacity, using capacity name to differentiate (e.g., `gpu-0-memory-block-0`)
* #27
* YAML for two DGXA100 nodes (each with eight A100 devices)
  * 341k
  * 12661 lines

**Option 2**
* Similar to Kevin's original proposal, as well as to [this](https://github.com/kubernetes/enhancements/pull/4709#discussion_r1639458926) 
* No attributes shared among devices
* Shared capacity for each physical GPU are grouped into named sets, rather than embedding the physical GPU name in the capacity name as in Option 1
* Consumption of those per device then name the specific groups from which they consume
* #34 
* YAML for two DGXA100 nodes (each with eight A100 devices)
  * 362k
  * 13525 lines

**Option 3**
* Like option 1, but share common attributes across all devices, overlaying device-specific ones
* #30 
* YAML for two DGXA100 nodes (each with eight A100 devices)
  * 205k
  * 7695 lines

**Option 4**
* Takes option 3 a step further, and defines a common shape for devices, not just common attributes
* In the non-partitioned case, this amounts to common attributes, so it also compacts things a lot in that case
* #31 
* YAML for two DGXA100 nodes (each with eight A100 devices)
  * 29k
[...]

PR #29, opened 2024-06-14
Paritionable model with grouped resources

johnbelamaric:
This is another alternative to the partitionable model.

The PR is a delta on top of #27.

Rather than flattening the individual physical GPU named resources into a single array, this is closer to Kevin's original, and keeps each physical GPU's shared resources in a separate struct. The individual devices then consume from those structs, by name. It is also possible for a device to consume from multiple shared resource groups, though that is not shown in the example.


dra-evolution/pkg/api/capacity_types.go
```diff
@@ -69,6 +78,25 @@ type ResourcePoolSpec struct {
...
+	Name string `json:"name"`
+
+	// SharedCapacity defines the sets of shared capacity consumable by
+	// devices in this SharedCapacityGroup.
+	//
+	// Must not have more than 16 entries.
+	//
+	// +listType=atomic
+	// +optional
+	SharedCapacity []SharedCapacity `json:"sharedCapacity,omitempty"`
```
klueska, line 96: Maybe `items` is a better name so it's not stuttering so much.
```suggestion
	Items []SharedCapacity `json:"items,omitempty"`
```
I'd also say that its not optional. If you have one of these, it should have at least one item in it.

dra-evolution/pkg/api/capacity_types.go
```diff
@@ -59,6 +59,15 @@ type ResourcePoolSpec struct {
...
 
+	// SharedCapacity defines the sets of shared capacity consumable by
+	// devices in this ResourceSlice.
+	//
+	// Must not have more than 16 entries.
+	//
+	// +listType=atomic
+	// +optional
+	SharedCapacity []SharedCapacityGroup `json:"sharedCapacity,omitempty"`
+
```
klueska, line 70: Does it make sense to double down on this PR and add the notion of `SharedAttributes` with a similar structure (i.e. a named group of attributes)? Then we can create groups of common attributes that individual devices can reference by name to include (in addition to their own local attributes, which would take precedence on nameing conflicts).

klueska: I had this in some proposal somewhere, but I an't seem to find it anymore.

johnbelamaric: That's combining this with option 3. I am working through various options, see https://github.com/kubernetes-sigs/wg-device-management/issues/20

Let's see all the options before we decide which to iterate on.

thockin:
Since this is "what's in the KEP" I think we should merge it and rebase all the options on it, so they appear as diffs.

/lgtm

thockin:
Shoot, I screwed up and approved the wrong one!  I meant to LGTM #27 !!!

PR #30, opened 2024-06-14
Partitionable with common attributes

johnbelamaric:
This is an evolution of the partitionable model defined in #27, which moves common attributes up to the pool level to reduce the size of the objects.

### DIFF
--- a/dra-evolution/pkg/api/capacity_types.go
+++ b/dra-evolution/pkg/api/capacity_types.go
@@ -68,6 +68,11 @@ type ResourcePoolSpec struct {
 	// +optional
 	SharedCapacity []SharedCapacity `json:"sharedCapacity,omitempty"`
 
+	// Attributes contains common device attributes that are the same
+	// for all devices in the pool, unless a device specifically over
+	// writes it by defining an attribute of the same name.
+	Attributes []DeviceAttribute `json:"attributes,omitempty"`
+
 	// Devices lists all available devices in this pool.
 	//
 	// Must not have more than 128 entries.
--- a/dra-evolution/testdata/pools-two-nodes-dgxa100.yaml
+++ b/dra-evolution/testdata/pools-two-nodes-dgxa100.yaml
@@ -4,22 +4,23 @@
     creationTimestamp: null
     name: nvidia-00-dgxa100
   spec:
+    attributes:
+    - name: brand
+      string: Nvidia
+    - name: architecture
+      string: Ampere
+    - name: cuda-compute-capability
+      version: 8.0.0
+    - name: driver-version
+      version: 550.54.15
+    - name: cuda-driver-version
+      version: 12.4.0
+    - name: product-name
+      string: Mock NVIDIA A100-SXM4-40GB
     devices:
     - attributes:
       - name: uuid
-        string: GPU-684cbd68-c82a-4100-9fde-e63109a9fac3
-      - name: product-name
-        string: Mock NVIDIA A100-SXM4-40GB
-      - name: brand
-        string: Nvidia
-      - name: architecture
-        string: Ampere
-      - name: cuda-compute-capability
-        version: 8.0.0
-      - name: driver-version
-        version: 550.54.15
-      - name: cuda-driver-version
-        version: 12.4.0
+        string: GPU-cdd0dd02-b5dd-4ae9-9c5a-c76b19086eb4
       - bool: true
         name: mig-capable
       - bool: false
@@ -31,18 +32,6 @@
     - attributes:
       - name: mig-profile
         string: 1g.5gb
-      - name: product-name
-        string: Mock NVIDIA A100-SXM4-40GB
-      - name: brand
-        string: Nvidia
-      - name: architecture
-        string: Ampere
-      - name: cuda-compute-capability
-        version: 8.0.0
-      - name: driver-version
-        version: 550.54.15
-      - name: cuda-driver-version
-        version: 12.4.0
       name: gpu-0-mig-1g.5gb-0
       sharedCapacityConsumed:
       - capacity: "14"
@@ -56,18 +45,6 @@
     - attributes:
       - name: mig-profile
         string: 1g.5gb
-      - name: product-name
-        string: Mock NVIDIA A100-SXM4-40GB
-      - name: brand
-        string: Nvidia
-      - name: architecture
-        string: Ampere
-      - name: cuda-compute-capability
-        version: 8.0.0
-      - name: driver-version
-        version: 550.54.15
-      - name: cuda-driver-version
-        version: 12.4.0
       name: gpu-0-mig-1g.5gb-1
       sharedCapacityConsumed:
       - capacity: "14"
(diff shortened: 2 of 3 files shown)


dra-evolution/pkg/api/capacity_types.go
```diff
@@ -68,6 +68,11 @@ type ResourcePoolSpec struct {
 	// +optional
 	SharedCapacity []SharedCapacity `json:"sharedCapacity,omitempty"`
 
+	// Attributes contains common device attributes that are the same
+	// for all devices in the pool, unless a device specifically over
```
pohly, line 72: "the pool" or "the slice"?

The prototype is not up-to-date here :cry: We should update it to match the current KEP before doing further prototyping.

pohly:
> to reduce the size of the objects

That helps reduce the size on average, but for the worst-case analysis which determines the limits of the slices we have to assume that the new `Attributes` is fully-populated *and* all devices have the maximum number of attributes.

dra-evolution/pkg/api/capacity_types.go
johnbelamaric: Yeah. I didn't want to do that yet, I figured it could be done and I could rebase...

johnbelamaric:
> > to reduce the size of the objects
> 
> That helps reduce the size on average, but for the worst-case analysis which determines the limits of the slices we have to assume that the new `Attributes` is fully-populated _and_ all devices have the maximum number of attributes.

Yes. True.

johnbelamaric:
I am making several options, see comment in #20

dra-evolution/testdata/pools-two-nodes-dgxa100.yaml
```diff
@@ -0,0 +1,7695 @@
...
+    - name: product-name
+      string: Mock NVIDIA A100-SXM4-40GB
+    devices:
+    - attributes:
+      - name: uuid
+        string: GPU-cdd0dd02-b5dd-4ae9-9c5a-c76b19086eb4
+      - bool: true
+        name: mig-capable
+      - bool: false
+        name: mig-enabled
```
thockin: NOTE: we should establish core guidelines - are attributes expected to be mutable or immutable?  Do we encourage drivers to embed status here or are attributes only *intrinsic* things about the device?  This one feels like status.

dra-evolution/testdata/pools-two-nodes-dgxa100.yaml
```diff
@@ -0,0 +1,7695 @@
...
+      - capacity: 4864Mi
+        name: gpu-0-memory
+      - capacity: "1"
+        name: gpu-0-memory-slices-0
+    - attributes:
+      - name: mig-profile
+        string: 1g.5gb+me
+      name: gpu-0-mig-1g.5gb-me-1
+      sharedCapacityConsumed:
+      - capacity: "14"
```
thockin, line 294: "capacity" is a producer term, not a consumer term.  This should be "quantity" or "amount" or "request" or something.

dra-evolution/testdata/pools-two-nodes-dgxa100.yaml
```diff
@@ -0,0 +1,7695 @@
...
+        name: gpu-0-copy-engines
+      - capacity: "1"
+        name: gpu-0-decoders
+      - capacity: "1"
+        name: gpu-0-jpeg-engines
+      - capacity: "1"
+        name: gpu-0-ofa-engines
+      - capacity: 4864Mi
+        name: gpu-0-memory
+      - capacity: "1"
```
thockin: MAYBE (really not sure) we should say that if `capacity` is omitted here, it means "all"?  Then it is sort of a fancy mutex for "partitions" and an quantity for fungibles.

Maybe not worthwhile

thockin (commented):
I think it is clear that shared attributes is an appropriate optimization to include.

I think this further emphasizes that some form of nesting would (IMO) make other optimizations possible.  There are attributes that apply to all "cards" (e.g. vendor).  There are some attributes that apply to all GPUs (MIG or not) on a single card (e.g. the card UUID, for matching).  There are some attributes that apply to each leaf device.  Why are we shoehorning that into 2 levels?

johnbelamaric:
Option 4 has some nesting. That's #31. It is much more efficient than this one.

johnbelamaric:
We could do more levels. Not clear the payoff is there.

johnbelamaric:
Option 3 makes common attributes. Option 4 makes common attributes AND common partition "map".

dra-evolution/testdata/pools-two-nodes-dgxa100.yaml
klueska: This should not be an attribute. MIG capable yes, enabled no. This is leftover from the structure I used to represent all of this in classic DRA. It is very compact, but also not generalizable.

dra-evolution/testdata/pools-two-nodes-dgxa100.yaml
klueska: It reuses the same struct as that used to declare the capacity. We would need a new struct then (which seems fine).

klueska:
I 100% agree we should have shared attributes. My first prototype had it, but at the time you said „let’s not prematurely optimize for size“ so we dropped it.

I’m still not sold on nesting though. I also had it originally (in my „recursive“ device model), but you all (rightly) convinced me to drop it. And now after working with flat devices and updating both the example driver and the NVIDIA GPU driver to  adhere to them, I’m really happy with the flexibility that a flat model brings us.

I really don’t think it buys us much to have nesting and I have a strong feeling will come back to bite us fairly quickly.

thockin:
I was continuing on the KEP PR, but detoured to these option, so I will copy a comment:

I just find it super weird to have gpu1's shared items consumable from gpu0.  This is the thing that is setting me on edge.  That implies to me a level of fungibility which doesn't exist.  There *is* a grouping that is smaller than a ResourceSlice but bigger than a Device, and we are not modelling it.  Call it a "card" for the moment.  A slice doesn't have shared resources, a "card" does.  Now you'll probably tell me "actually, one card can borrow resources on another card".  In fact, I can already see the (hypothetical) use-case for a channelized `<something>` which can be effectively RAID'ed into a larger logical device.  But that's not this, and (TTBOMK) that doesn't exist yet.

I really can see both sides, and I don't mean to be dogmatic.  It just smells funny.  Let's keep the conversation overall moving forward, and if this is all that's left, we can hash it out.

>  My first prototype had it, but at the time you said „let’s not prematurely optimize for size“ so we dropped it.

Yeah, we dropped a LOT to get the baseline, and bringing partitions back makes it clear to me that this is one piece that really does make sense.

dra-evolution/testdata/pools-two-nodes-dgxa100.yaml
```diff
@@ -0,0 +1,7695 @@
...
+      - bool: true
+        name: mig-capable
+      - bool: false
+        name: mig-enabled
+      name: gpu-0
+      sharedCapacityConsumed:
+      - capacity: 40Gi
+        name: gpu-0-memory
+    - attributes:
+      - name: mig-profile
```
thockin, line 33: We should include in this YAML the requirement that all MIG partitions on a "card" have a "parent UUID" attribute with the same value (for matching).  I think that would highlight the "hidden" nesting level, even if we ultimately choose not to model it.

PR #32, opened 2024-06-14
Partitionable model with generated partitions

johnbelamaric:
Ok, this is option 5 in #20, and it build on top of option 4 which is in #31.

It's a little bit crazy. Essentially, instead of enumerating each partition, we use CEL expressions to allow algorithmic creation of the partitions. For example, there are 7 migs with the profile `1g.5gb`. They only really differ in their name and which specific memory slices they consume. So, instead of listing all 7, we create a "format" such that iterating over it from 0 to 6 we can instantiate the 7 partitions.

This should work, it is more compact even with the current number of memory slices. If there were 100 memory slices, this would make a huge difference compared to #31.


thockin (commented):
Than you for finding "the line" and showing us what happens if we cross it :)

I'm -2 on this one :)

PR #36, opened 2024-06-18
Update API for 1.31 without partitionable model

johnbelamaric:
This attempts to update the API to:

1) Remove partitionable model
2) Allow extension points for future addition of partitionable model, in line with the discussions in #20.

This is rough, not exactly what would go into k/k. But I think it has what we need. It allows us to do common attrs and common capacity (*common*, not *shared*). It looks little like some of Kevin's earlier prototypes and a sort of simplified version of #31 and #35.

cc @pohly @klueska

### DIFF
--- a/dra-evolution/pkg/api/capacity_types.go
+++ b/dra-evolution/pkg/api/capacity_types.go
@@ -92,9 +92,6 @@ type Device struct {
 	//
 	// Must not have more than 32 entries.
 	//
-	// Values in this list whose name conflict with common attributes
-	// will override the common attribute values.
-	//
 	// +listType=atomic
 	// +optional
 	//
@@ -104,9 +101,6 @@ type Device struct {
 	// The name of each capacity must be unique.
 	//
 	// Must not have more than 32 entries.
-	//
-	// Values in this list whose name conflict with common capacity
-	// will override the common capacity values.
 	//
 	// +listType=atomic
 	// +optional
--- a/dra-evolution/pkg/api/capacity_types.go
+++ b/dra-evolution/pkg/api/capacity_types.go
@@ -59,14 +59,11 @@ type ResourcePoolSpec struct {
 	// vendor of the driver.
 	DriverName string `json:"driverName" protobuf:"bytes,3,name=driverName"`
 
-	// SharedCapacity defines the set of shared capacity consumable by
-	// devices in this ResourceSlice.
+	// CommonData represents information that is common across various devices,
+	// to reduce redundancy in the devices list.
 	//
-	// Must not have more than 128 entries.
-	//
-	// +listType=atomic
 	// +optional
-	SharedCapacity []SharedCapacity `json:"sharedCapacity,omitempty"`
+	CommonData []CommonDeviceData `json:"commonData,omitempty"`
 
 	// Devices lists all available devices in this pool.
 	//
@@ -78,8 +75,25 @@ type ResourcePoolSpec struct {
 	// them) empty pool.
 }
 
-const ResourcePoolMaxSharedCapacity = 128
+const ResourcePoolMaxCommonData = 16
 const ResourcePoolMaxDevices = 128
+
+type CommonDeviceData struct {
+	// Name identifies the particular set of common data represented
+	// here.
+	//
+	// +required
+	Name string `json:"name"`
+
+	// Exactly one must be populated
+	Attributes []DeviceAttribute
+	Capacity   []DeviceCapacity
+
+	// Future possible additions:
+	// - SharedCapacity for use in allocating partitionable devices
+	// - DeviceShape for encapsulating attributes, capacities, partion schemes
+	// - DeviceShapeRef for storing those in a secondary object
+}
 
 // Device represents one individual hardware instance that can be selected based
 // on its attributes.
@@ -88,27 +102,43 @@ type Device struct {
 	// the driver on the node. It must be a DNS label.
 	Name string `json:"name" protobuf:"bytes,1,name=name"`
 
+	// CommonData is the list of names of entries from the common data that
+	// should be included in this device.
+	//
+	// +listType=atomic
+	// +optional
+	//
+	CommonData []string `json:"commonData,omitempty"`
+
 	// Attributes defines the attributes of this device.
 	// The name of each attribute must be unique.
 	//
 	// Must not have more than 32 entries.
 	//
+	// Values in this list whose name conflict with common attributes
+	// will override the common attribute values.
+	//
 	// +listType=atomic
 	// +optional
+	//
 	Attributes []DeviceAttribute `json:"attributes,omitempty" protobuf:"bytes,3,opt,name=attributes"`
 
-	// SharedCapacityConsumed defines the set of shared capacity consumed by
-	// this device.
+	// Capacity defines the capacity values for this device.
+	// The name of each capacity must be unique.
 	//
 	// Must not have more than 32 entries.
 	//
+	// Values in this list whose name conflict with common capacity
+	// will override the common capacity values.
+	//
 	// +listType=atomic
 	// +optional
-	SharedCapacityConsumed []SharedCapacity `json:"sharedCapacityConsumed,omitempty"`
+	//
+	Capacity []DeviceCapacity `json:"capacity,omitempty"`
 }
 
 const ResourcePoolMaxAttributesPerDevice = 32
-const ResourcePoolMaxSharedCapacityConsumedPerDevice = 32
+const ResourcePoolMaxCapacityPerDevice = 32
 
 // ResourcePoolMaxDevices and ResourcePoolMaxAttributesPerDevice where chosen
 // so that with the maximum attribute length of 96 characters the total size of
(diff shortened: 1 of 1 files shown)


dra-evolution/pkg/api/capacity_types.go
```diff
@@ -78,37 +75,70 @@ type ResourcePoolSpec struct {
...
 type Device struct {
 	// Name is unique identifier among all devices managed by
 	// the driver on the node. It must be a DNS label.
 	Name string `json:"name" protobuf:"bytes,1,name=name"`
 
+	// CommonData is the list of names of entries from the common data that
+	// should be included in this device.
+	//
+	// +listType=atomic
+	// +optional
```
johnbelamaric: Also, "last in wins". So if I list mulitiple "common attribute sets", they are overlayed in the order listed (with the Device's attributes last).

dra-evolution/pkg/api/capacity_types.go
```diff
@@ -59,14 +59,11 @@ type ResourcePoolSpec struct {
...
-	// devices in this ResourceSlice.
+	// CommonData represents information that is common across various devices,
+	// to reduce redundancy in the devices list.
 	//
-	// Must not have more than 128 entries.
-	//
-	// +listType=atomic
 	// +optional
-	SharedCapacity []SharedCapacity `json:"sharedCapacity,omitempty"`
+	CommonData []CommonDeviceData `json:"commonData,omitempty"`
```
klueska: I wouldn't include this. If we are ripping out the model that supports partitioning, I think we should should just go back to an explicit list of attributes per device.

dra-evolution/pkg/api/capacity_types.go
```diff
@@ -154,40 +184,24 @@ type DeviceAttribute struct {
...
-	//
 	// +required
 	Name string `json:"name"`
 
 	// Capacity is the total capacity of the named resource.
-	// This can either represent the total *available* capacity, or the total
-	// capacity *consumed*, depending on the context where it is referenced.
 	//
 	// +required
 	Capacity resource.Quantity `json:"capacity"`
```
klueska, line 166: I would make this. one-of for local capacity (now) and shared capacity (later)

dra-evolution/pkg/api/capacity_types.go
```diff
@@ -78,37 +75,70 @@ type ResourcePoolSpec struct {
...
 	// the driver on the node. It must be a DNS label.
 	Name string `json:"name" protobuf:"bytes,1,name=name"`
 
+	// CommonData is the list of names of entries from the common data that
+	// should be included in this device.
+	//
+	// +listType=atomic
+	// +optional
+	//
+	CommonData []string `json:"commonData,omitempty"`
```
klueska: As as mentioned above -- I would remove this until we know how we are going to support partitioning.

dra-evolution/pkg/api/capacity_types.go
johnbelamaric: We need an extension point to insert the things we are debating in #20. That's the primary reason that is here. Adding the common things was just a way to make it have some meaning and not be a stub. With this, an old scheduler an see that there is an entry in this list, but when it unmarshals it, all the fields the scheduler is aware of are nil, so it knows "ooo, I should ignore this slice, I don't know what to do with it".

If we don't have this, then we need some other way for an older scheduler to know that "it doesn't understand this object".

johnbelamaric: The alternative would be to have a discriminator field, like `SliceContent: "SimpleDevices"` or something. Then we would add new values to that so old schedulers know if they should handle the slice or not.

dra-evolution/pkg/api/capacity_types.go
johnbelamaric: In this version, that is handled by using a the `CommonDeviceData` one-of (`Capacity` and `SharedCapacity` in that one-of). We could also add a separate, `SharedCapacity` field at the device level if we wanted individual devices to offer shared capacity (depends on how we end up doing partitions whether that makes any sense or not).

dra-evolution/pkg/api/capacity_types.go
klueska: My current thinking is that we would *not* have any extension points directly in the device at this point. Instead every device will  have the ability to point to one (and exactly one) DeviceShape/DeviceTemplate. That is what will have all of these extra bells and whistles. The device itself is still just a named list of attributes (and now capacities).

johnbelamaric: Ok, that works for me. The thing is, we still need a way for old schedulers to know "this slice is not something I understand". So a "SliceType" discriminator perhaps? Or we have to have a field like I have done here, that has a one-of.

dra-evolution/pkg/api/capacity_types.go
johnbelamaric: I don't know that we want to model it exactly that way, and it won't come into play until we have a different discriminator value, so I don't think it's necessary.

dra-evolution/pkg/api/capacity_types.go
```diff
@@ -59,14 +59,13 @@ type ResourcePoolSpec struct {
...
+	// to this discriminator. This allows older schedulers to know if they should
+	// ignore the slice, due to it being of a type they do not understand.
 	//
-	// Must not have more than 128 entries.
-	//
-	// +listType=atomic
-	// +optional
-	SharedCapacity []SharedCapacity `json:"sharedCapacity,omitempty"`
+	// +required
+	SliceContent string `json:"sliceContent"`
```
johnbelamaric, line 68: Should this be an enum instead, @pohly ?

johnbelamaric:
Ok, I simplifed it and now it's basically just a discriminator with a single valid value in 1.31, but more will come.

dra-evolution/pkg/api/capacity_types.go
```diff
@@ -93,22 +92,30 @@ type Device struct {
 	//
 	// Must not have more than 32 entries.
 	//
+	// Values in this list whose name conflict with common attributes
+	// will override the common attribute values.
+	//
```
klueska: This comment is no longer needed.

dra-evolution/pkg/api/capacity_types.go
```diff
@@ -93,22 +92,30 @@ type Device struct {
...
-	// SharedCapacityConsumed defines the set of shared capacity consumed by
-	// this device.
+	// Capacity defines the capacity values for this device.
+	// The name of each capacity must be unique.
 	//
 	// Must not have more than 32 entries.
 	//
+	// Values in this list whose name conflict with common capacity
+	// will override the common capacity values.
+	//
```
klueska: This comment is no longer needed.

dra-evolution/pkg/api/capacity_types.go
klueska: Sure. If we are not in agreement, then what is there now seems fine. As you said, we can have a separate SharedCapacity slice in the future to cover this (though I'm probably going to argue we don't need it).

dra-evolution/pkg/api/capacity_types.go
klueska: So the scheduler will decode this object, read this field, and then decide if it knows how to work with it or not? an enum seems appropriate then, since this should only be allowed to have a specific set of values.

pohly: Do we need this?

If a current-rev scheduler only knows how to work with "devices", it uses the `slice.devices` field to find those.

If a future API adds "gizmos" or some variant of "devices" (remember, I wanted to keep the current field named as "namedDevices"...), then it adds a `slice.gizmos` field. A current-rev scheduler will see an empty slice and ignore it because it contains no devices.

pohly: Or is the intent to add additional fields to the `Device` struct? In that case such a `SliceContent` discriminator would be useful because an unknown value tells old clients that there is something in the `slice.devices` that they don't know about, but should know.

I just wouldn't call it `SliceContent`, but rather `DeviceType`: it's a description of the `slice.devices`.

If we add `slice.gizmos` (separate field), then they can have a `slice.gizmoType`.

johnbelamaric: We likely would add another field in addition to "Devices" to qualify or add onto the devices. 

If we want to avoid the discriminator, I would avoid the term "Devices". But I don't think "NamedDevices" works because "named" isn't a differentiating feature. Maybe "SimpleDevices" or something?

If we keep the discriminatory, I do not think "DeviceType" is right - because it wouldn't make sense for "gizmos", for example.

klueska: I like the discriminator and don't feel strongly about what it is called (though I'd like it to have some symmetry with other fields -- what did we decide to call the replacement for `PoolDeviceCount` now that we are going with a `SliceCount` instead)?

I think Patrick is suggesting that field itself becomes a one-of (with a future `GizmoType` field, for example), such that it pairs with the inclusion of a `devices` list (or a `gizmos` list in the future).

pohly: My point was that there is no need for one-ofs. A slice could contain both `SimpleDevices []SimpleDevice` and `ComplexDevices []ComplexDevice` at the same time. Depending what a scheduler supports, it can allocate from `SimpleDevices` or from both.

However, for the maximum size calculation it might make sense to require that either one or the other slice is empty. That doesn't change my argument, a scheduler which only supports `SimpleDevices []SimpleDevice` will see an empty slice and safely ignore it.

pohly: Having said that, I am okay with adding the discriminator. It may be redundant when adding a second slice with a different type, but for the case where the same type gets extended (https://github.com/kubernetes-sigs/wg-device-management/pull/36/files/3a7aa4f9789c7979426ea3c65497545136428f55#r1646662809) it is useful.

My question about validation after a downgrade applies also here.

johnbelamaric: According to the bullet on "enumerated types" [here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#backward-compatibility-gotchas), we should be OK as long as we document that schedulers that see unknown slice types should ignore those slices. I think older clients doing a *write* would be subject to the API validation of the API they are using, meaning they would not be able to use the new enum. I think if there is a field associated with the new enum value, we should clear it (the client doesn't know about the new enum, but we do). I *think* that's consistent with the logic described [here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#new-field-in-existing-api-version), but we should discuss with api reviewers. 

One question: Should (can?) we ensure all slices of a pool are of the same slice type? If not, old schedulers would have to ignore the whole pool, at least for "all", if ANY slice is of an unknown type...

johnbelamaric: It would seem to me that adding an enum value should require a bump in the API version.

pohly: Once we have reached v1, we cannot bump it any further?

I'm still not exactly clear on how to implement validation, but there must have been precedences, so I guess it is okay.

thockin: A discriminator is *required* when you want the absence of config to imply a default, so that version skew isn't perceived as the default case.

For cases that are truly one-of (exactly one field, empty lists don't count), then a discriminator is not needed, because any actor can perceive the lack of a value, and choose not to act.  A discriminator can make for better error messages, but that's about all.

Whether we extend `Device` to encompass partitions (assuming we can do that safely) or we add `PartitionableDevice` here, it shouldn't matter -- both are viable.  Likewise, whether we want devices/gizmos/widgets/partitionables to co-exist in a slice or to be mutually exclusive, I think both are viable.  There may be *other* things to consider to make those decisions, but I don't think *this* aspect matters.

PR #40, opened 2025-08-17
create a page for tracking device ecosystem integration

kannon92:
Draft of https://github.com/kubernetes/website/issues/51808.

In that issue, it seems there is not interest in having k/website track third party integrations.

For now, I wanted to have a place where we can update status of DRA ecosystem.

### DIFF
--- a/device-ecosystem/README.md
+++ b/device-ecosystem/README.md
@@ -2,22 +2,24 @@
 
 DRA is going to stable in 1.34 of Kubernetes. As more device vendors provide support for DRA, we want to have a tracking page where people can do to see who supports DRA.
 
-| Vendor  | DRA Driver
-|---------|------------
-| NVIDIA  | [dra-driver](https://github.com/NVIDIA/k8s-dra-driver-gpu)
-| AMD     | could not find
-| Intel   | [intel-resource-drivers-for-kubernetes](https://github.com/intel/intel-resource-drivers-for-kubernetes)
-| Google (TPU)  | could not find
-
 ## Contributing
 
 We expect the status of the DRA drivers and their corresponds operators to change over time. Please feel free to add implementations as they graduate.
+
+This document is not official supported by Kubernetes and is mainly used by #wg-device-management to track where device vendors are in their integrations for DRA.
 
 ## DRA Features
 
 One area that we want to track is not just DRA availability but also support of various DRA features as they graduate.
 
 ### Table
+
+| Vendor  | DRA Driver
+|---------|------------
+| NVIDIA  | [dra-driver](https://github.com/NVIDIA/k8s-dra-driver-gpu)
+| AMD     | could not find
+| Intel   | [intel-resource-drivers-for-kubernetes](https://github.com/intel/intel-resource-drivers-for-kubernetes)
+| Google (TPU)  | could not find
 
 
 ### NVIDIA
--- a/device-ecosystem/README.md
+++ b/device-ecosystem/README.md
@@ -1,0 +1,55 @@
+# Device Ecosystem for DRA
+
+DRA is going to stable in 1.34 of Kubernetes. As more device vendors provide support for DRA, we want to have a tracking page where people can do to see who supports DRA.
+
+| Vendor  | DRA Driver
+|---------|------------
+| NVIDIA  | [dra-driver](https://github.com/NVIDIA/k8s-dra-driver-gpu)
+| AMD     | could not find
+| Intel   | [intel-resource-drivers-for-kubernetes](https://github.com/intel/intel-resource-drivers-for-kubernetes)
+| Google (TPU)  | could not find
+
+## Contributing
+
+We expect the status of the DRA drivers and their corresponds operators to change over time. Please feel free to add implementations as they graduate.
+
+## DRA Features
+
+One area that we want to track is not just DRA availability but also support of various DRA features as they graduate.
+
+### Table
+
+
+### NVIDIA
+
+#### NVIDIA DRA Driver
+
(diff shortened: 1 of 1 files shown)


kannon92:
cc @johnbelamaric @pohly 

I'm not entirely sure where this work should live right now. I figured we can at least have this here for now and eventually move somewhere more official if we want.

lmktfy (changes requested):
CNCF policy recommends alphabetical order (either by vendor or product). Alternatively, use a stable random order, eg assign a sort value each time you add a vendor.

device-ecosystem/README.md
```diff
@@ -0,0 +1,55 @@
+# Device Ecosystem for DRA
+
+DRA is going to stable in 1.34 of Kubernetes. As more device vendors provide support for DRA, we want to have a tracking page where people can do to see who supports DRA.
```
lmktfy, line 3: This will soon go stale

device-ecosystem/README.md
```diff
@@ -0,0 +1,55 @@
+# Device Ecosystem for DRA
+
```
lmktfy, line 2: Can we add that this page is not formally maintained?

lmktfy:
> In that issue, it seems there is not interest in having k/website track third party integrations.

Not quite. I think we (I) said that it can go in k/website provided it lives in the blog section.

You can still do that. And you can either update the article as you wish, or post a new article whilst making the old one link to the new one.

kannon92:
> > In that issue, it seems there is not interest in having k/website track third party integrations.
> 
> Not quite. I think we (I) said that it can go in k/website provided it lives in the blog section.
> 
> You can still do that. And you can either update the article as you wish, or post a new article whilst making the old one link to the new one.

I don't think a blog is what we want here. I wanted a place to go to get updates on device vendors support for DRA.

I'm finding it difficult to keep together all the open source milestones for GPU vendors. I expect this to change over time and I want to update this doc as new DRA features are released and device vendors may take time to support them.

So I still don't think a blog is the right place for this.

device-ecosystem/README.md
kannon92: Sure!

guptaNswati: This seems like a good place to host all the vendor drivers for DRA. If this does not get accepted, we may also add a section to example-dra-driver.

device-ecosystem/README.md
```diff
@@ -0,0 +1,57 @@
...
+We expect the status of the DRA drivers and their corresponds operators to change over time. Please feel free to add implementations as they graduate.
+
+This document is not official supported by Kubernetes and is mainly used by #wg-device-management to track where device vendors are in their integrations for DRA.
+
+## DRA Features
+
+One area that we want to track is not just DRA availability but also support of various DRA features as they graduate.
+
+### Table
+
```
guptaNswati, line 16: may be also list the example-dra-driver

device-ecosystem/README.md
```diff
@@ -0,0 +1,57 @@
...
+| Vendor  | DRA Driver
+|---------|------------
```
guptaNswati, line 18: The format can be vendor driver description with their landing page, list of supported features like a checklist of what is available from the upstream and supported hardware types. I can do an example entry for NVIDIA drivers

device-ecosystem/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+### NVIDIA
+
+#### NVIDIA DRA Driver
+
+NVIDIA has the [dra-driver](https://github.com/NVIDIA/k8s-dra-driver-gpu).
+
+As of August 2025, one can install the DRA-driver via a helm chart to support NVIDIA + DRA.
+
+#### GPU Operator
```
guptaNswati, line 33: mmm, not sure if it needs a separate entry. It can be part of NVIDIA-dra driver as an option to install them as part of gpu-operator

device-ecosystem/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+Searching in [AMD ROCm](https://github.com/ROCm) I cannot seem to find a DRA implementation yet.
+
+TODO: maybe some help on AMD on status on this.
+
+### Google
+
+#### Google TPU Driver
+
+I wasn't able to find this. TODO: maybe some help on google on status on this.
```
johnbelamaric, line 57: Since TPUs are only in GCE, we haven't open sourced this at this point. But we can provide a link to the Google Cloud docs page when it's ready. cc @mortent 

We should have an entry for dranet. Currently it's Google OSS, but...

johnbelamaric: cc @LionelJouin @aojea 

Also possibly the CNI DRA driver

johnbelamaric: and of course the CPU driver cc @pravk03 (k8s project / k-sigs owned)

kannon92: Is there a possibility to merge this as a draft? And we can collbororate on this.

device-ecosystem/README.md
kannon92: Yea I guess this is my Openshift bias as we are really looking to support GPU use cases via GPU Operator.

I can remove the operator but I like bundling is important as this may be of interest to downstream distributions.

So even if DRA driver is provided we would want some way to bundle it without having to require admins to install their own chart. But t

device-ecosystem/README.md
kannon92: Yea, I'm not sure the path forward here.

This work will always be changing and we do need some place to write this information down.

johnbelamaric:
/approve
/lgtm

Merging to allow subsequent PRs to update it from other people

